PR #36, opened 2019-01-02
Implement String.prototype's toLowerCase and toUpperCase

targos:
Refs: https://github.com/devsnek/engine262/issues/22


targos:
Landed with https://github.com/devsnek/engine262/pull/50

PR #64, opened 2019-01-13
MakeDay

Yaffle:


src/abstract-ops/date-objects.mjs
```diff
@@ -201,18 +201,22 @@ export function MakeTime(hour, min, sec, ms) {
...
+   if (!Number.isFinite(year.numberValue()) || !Number.isFinite(month.numberValue()) || !Number.isFinite(date.numberValue())) {
+     return new Value(NaN);
+   }
+   const y = X(ToInteger(year)).numberValue();
+   const m = X(ToInteger(month)).numberValue();
+   const dt = X(ToInteger(date)).numberValue();
+   const ym = y + Math.floor(m / 12);
+   const mn = mod(m, 12);
+//   TODO: 7. Find a value t such that YearFromTime(t) is ym and MonthFromTime(t) is mn and DateFromTime(t) is 1; but if this is not possible (because some argument is out of range), return NaN.
+   const ymday = DayFromYear(new Value(ym + (mn > 1 ? 1 : 0))).numberValue() - 365 * (mn > 1 ? 1 : 0) + daysWithinYearToEndOfMonth[mn];
```
TimothyGu: 1. This doesn’t check if the day belongs to the month (what if `dt` is -1 or 32?).
2. A negative `mn` here is possible, which doesn’t sound like what we want.
3. Are leap years accounted for?
4. `t` may overflow `Number.MAX_SAFE_INTEGER` (or underflow `Number.MIN_SAFE_INTEGER`). We should return `NaN` if it’s outside of that range.
5. I don’t quite understand `mn > 1 ? 1 : 0`… can you explain?

targos:
I fixed the indentation and the function (commented code had a bug) and merged.
This is good enough to enable almost all Date tests. Let's fix edge cases separately.

Yaffle:
Thank you!
Well... as @TimothyGu  pointed out, `MakeDay` will not work when some values are passed because of accuracy errors. The spec is not clear here. Even the step `Let ym be y + floor(m / 12)` cannot be done without BigInt arithmetic. But the specification is saying to return `NaN` when the result cannobe be computed only at step `7`.
So, yes, `MakeDay` still needs to be updated.

PR #70, opened 2019-01-24
[WIP] RegExp grammar with nearley

targos:
/cc @TimothyGu I'll probably need your help with this.


src/grammar/RegExp.ne
```diff
@@ -0,0 +1,219 @@
...
+
+GroupName ->
+    "<" RegExpIdentifierName ">"
+
+RegExpIdentifierName ->
+    RegExpIdentifierStart 
+  | RegExpIdentifierName RegExpIdentifierPart 
+
+RegExpIdentifierStart ->
+    [a-zA-Z]# UnicodeIDStart
```
targos: I have no idea how to encode `UnicodeIDStart` (and `UnicodeIDContinue` on line 93)

TimothyGu:
Hmm, the negative lookaheads are problematic. I don’t know if nearley actually supports those.

TimothyGu:
Just going to put that here: https://github.com/acornjs/acorn/blob/master/acorn/src/regexp.js looks like a good parser, though doesn’t look like it supplies an AST.

TimothyGu:
Let's transfer your branch to devsnek/engine262: #71.

PR #100, opened 2020-04-13
implement String.prototype.replaceAll proposal

targos:
Refs: https://tc39.es/proposal-string-replaceall/

### DIFF
--- a/src/intrinsics/StringPrototype.mjs
+++ b/src/intrinsics/StringPrototype.mjs
@@ -23,7 +23,12 @@ import {
   Value,
   wellKnownSymbols,
 } from '../value.mjs';
-import { GetSubstitution, TrimString, StringPad } from '../runtime-semantics/all.mjs';
+import {
+  GetSubstitution,
+  StringIndexOf,
+  StringPad,
+  TrimString,
+} from '../runtime-semantics/all.mjs';
 import { Q, X } from '../completion.mjs';
 import { CreateStringIterator } from './StringIteratorPrototype.mjs';
 import { assignProps } from './Bootstrap.mjs';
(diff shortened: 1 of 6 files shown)


src/intrinsics/StringPrototype.mjs
```diff
@@ -154,29 +159,49 @@ function StringProto_includes([searchString = Value.undefined, position = Value.
...
-      if (searchStr[j] !== S[k + j]) {
-        match = false;
-        break;
+  // 8. Let searchLen be the length of searchStr.
+  const searchLen = searchStr.stringValue().length;
+  // https://tc39.es/proposal-string-replaceall/#sec-string.prototype.indexof
+  if (surroundingAgent.feature('String.prototype.replaceAll')) {
+    // Let position be ! StringIndexOf(S, searchString, start).
+    // TODO: fix bug in spect text.
+    position = X(StringIndexOf(new Value(S), searchStr, start));
```
targos, line 182: /cc @mathiasbynens I think that's a bug in the spec.

targos:
Added comments and fixed a spec bug. I forgot to run the whole test suite the first time.

src/intrinsics/StringPrototype.mjs
mathiasbynens: Sorry, I lack context here; what's the bug? Would you mind submitting a patch against https://github.com/tc39/proposal-string-replaceall?

mathiasbynens: The relevant line is here: https://github.com/tc39/proposal-string-replaceall/blob/33f5e50dd27ade44a52a586933d8ec32bbee164b/spec.html#L131

targos: I opened an issue: https://github.com/tc39/proposal-string-replaceall/issues/36

mathiasbynens: I see. Good catch! Thank you for filing the bug :)

PR #111, opened 2020-07-25
RegExp: Implement the legacy RegExp features proposal

ExE-Boss:
**Spec:** <https://github.com/tc39/proposal-regexp-legacy-features>
**Tests:** <https://github.com/tc39/test262/pull/2650>


ExE-Boss:
@devsnek Moved to: <https://github.com/engine262/engine262/pull/114>

PR #157, opened 2021-02-17
Error: Store stack frames in ErrorData

ExE-Boss:
This partially implements the **Error Stacks** proposal, specifically <https://github.com/tc39/proposal-error-stacks/pull/36>, <https://github.com/tc39/proposal-error-stacks/pull/37> and <https://github.com/tc39/proposal-error-stacks/pull/38>.

This also makes the `stack` property an accessor on the `%Error.prototype%` object.

## Depends on:
- <https://github.com/engine262/engine262/pull/259>


Jack-Works:
close since https://github.com/engine262/engine262/pull/259 is merged

PR #189, opened 2021-12-24
fix: isIntegerIndex should check the argument wether is an integer

ahabhgk:
According to the spec the `integer index` should be an integer ([spec](https://tc39.es/ecma262/#integer-index)), but the original code didn't check it.

### DIFF
--- a/src/abstract-ops/data-types-and-values.mjs
+++ b/src/abstract-ops/data-types-and-values.mjs
@@ -14,6 +14,9 @@ export function isIntegerIndex(V) {
   if (numeric === Value.undefined) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


ExE-Boss:
It does check it in the call to <code>[Number.isSafeInteger]\(…\)</code>: <https://github.com/engine262/engine262/blob/253bed7f19137f154cbe1f57ccb24edbbce9edf4/src/abstract-ops/data-types-and-values.mjs#L20>

[Number.isSafeInteger]: https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger

ahabhgk:
Oh.. Sorry for didn't notice it.

PR #203, opened 2022-12-05
Preview: add jsdoc-style type comment to some files

Jack-Works:
this is a preview of #202. using jsdoc-style type comment to use the type checker. as you can see, the type comment is very verbose. Do you want to go with JSDoc or with real TypeScript file?

This won't be the final PR to be merged. I intended to send separate PRs to do some changes before migrating some code (please comment if you don't like any of the changes listed below so I can avoid creating a PR that won't be accepted):

- convert comment `// #sec-something` to `/** https://tc39.es/ecma262/#sec-something */`
- Replace `Type(x) === 'Object'` with `x instanceof ObjectValue`.
      This reflects to the spec convention change from `if Type(x) is *Object*` to `if x is an Object`. and it's also more friendly to the TypeScript checker.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 dist
+declaration
 test.mjs
 test.js
 .eslintcache
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,7 @@
 test
 src
+declaration/.tsbuildinfo
+declaration/**/*.map
 scripts
 coverage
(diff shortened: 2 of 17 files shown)


Jack-Works:
Cool. Since engine262 is currently using babel Marco to support ? and ! (already cannot be executed directly) I think it's ok to introduce typescript. I'll open another pr to do that

PR #206, opened 2022-12-10
refactor: convert parser and some other file to typescript

Jack-Works:
meta: #202

Note: Commit cc2f5859b1524533693f5c321f8b117c65a85cf6 should appear in the main history after merge. I use a [`.git-blame-ignore-revs`](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/) file to ignore the automatic change in this commit.

# Infra changes

Add TypeScript support in eslint and rollup config. Does not add type checker to the CI yet.

## Rename all files

1. Rename all files from .mjs to .mts
2. Add `// @ts-nocheck` at the top of the file

This allows me to write TypeScript syntax in any file but not be type checked.
When any file is ready for type checked, just remove the `// @ts-nocheck` comment.

## Babel target

Chrome 70 supports optional chaining, and the previous config transforms it.

I need to introduce @babel/preset-env to use class field, ...etc syntaxes. I set the target to Chrome 69, but maybe we can get newer?

## Class fields

TypeScript requires class fields syntax to declare a field. Using [[Define]].

This should be a breaking change for downstream users (if there is any).

## Decorator

I use decorators to create callable classes.

```js
export @callable() class Descriptor { ... }
```

## Shipping declarations

Now after running `tsc -p .`, declarations/ folder contains the `.d.ts` file of engine262

## Remove @engine262/valid-throw rule

The throw() is now fully typed. An invalid throw message will be a type error.

# API changes

## Deprecates `new Value(val)` in flavor of `Value.of(val)`

TypeScript does not allow overload for class constructors so I cannot type this usage. I also marked `Value` as an abstract class therefore its class constructor is not allowed to be called in TypeScript.


rollup.config.js
```diff
@@ -18,17 +19,36 @@ const banner = `/*!
...
   plugins: [
     json({ compact: true }),
     commonjs(),
+    mtsResolver(),
     nodeResolve(),
     babel({
       babelHelpers: 'bundled',
       exclude: 'node_modules/**',
+      presets: [[
+        '@babel/preset-env',
```
devsnek, line 32: i don't think we need to do babel envs and such here. if someone wants to run this on old safari they can babel it themselves. one of the goals of this package is that the built output is as similar to the source as possible so that source map errors are as small as possible so that you can easily use js debuggers to step through it.

devsnek:
until all files have ts enabled, we should probably keep the valid throw eslint rule right?

src/value.mts
```diff
@@ -720,77 +750,86 @@ export class ObjectValue extends Value {
...
+  readonly Configurable?: BooleanValue;
+  constructor(O: Pick<Descriptor, 'Configurable' | 'Enumerable' | 'Get' | 'Set' | 'Value' | 'Writable'>) {
+    this.Value = O.Value;
+    this.Get = O.Get;
+    this.Set = O.Set;
+    this.Writable = O.Writable;
+    this.Enumerable = O.Enumerable;
+    this.Configurable = O.Configurable;
+  }
+  everyFieldIsAbsent() {
```
devsnek, line 805: whatever eslint rule is missing that ensures there would be newlines between class members, could you please enable it

src/value.mts
```diff
@@ -720,77 +750,86 @@ export class ObjectValue extends Value {
...
+      super(sizeOrBuffer, byteOffset, length);
     } else {
       Assert(typeof sizeOrBuffer === 'number');
       super(sizeOrBuffer);
     }
   }
 }
 
-export function Type(val) {
+export function Type(val: Value) {
```
devsnek, line 832: this annotation is not correct?

rollup.config.js
```diff
@@ -56,3 +76,21 @@ module.exports = () => ({
     warn(warning);
   },
 });
+function mtsResolver() {
```
devsnek, line 79: just update all the imports?

src/parser/FunctionParser.mts
```diff
@@ -1,14 +1,19 @@
...
+import { ArrowInfoStack, getDeclarations } from './Scope.mjs';
 import { Token } from './tokens.mjs';
 import { IdentifierParser } from './IdentifierParser.mjs';
+import type { ParseNode } from './Parser.mjs';
 
-export const FunctionKind = {
-  NORMAL: 0,
-  ASYNC: 1,
-};
+export enum FunctionKind {
```
devsnek, line 7: what does this compile to? will it be missing in the public api now?

src/parser/FunctionParser.mts
```diff
@@ -222,18 +227,18 @@ export class FunctionParser extends IdentifierParser {
     }
   }
 
-  parseArrowFunction(node, { arrowInfo, Arguments }, kind) {
+  parseArrowFunction(node: ParseNode, { arrowInfo, Arguments }: ParseNode | { arrowInfo?: ArrowInfoStack, Arguments: readonly ParseNode[] }, kind: FunctionKind) {
     const isAsync = kind === FunctionKind.ASYNC;
     this.expect(Token.ARROW);
     if (arrowInfo) {
-      arrowInfo.awaitExpressions.forEach((e) => {
+      arrowInfo.awaitExpressions.forEach((e: ParseNode) => {
```
devsnek, line 234: are all these type annotations needed? `arrowInfo.whatever` should be typed already right?

src/parser/RegExpParser.mts
```diff
@@ -918,3 +924,116 @@ export class RegExpParser {
     };
   }
 }
+export interface QuantifierNodePrefix {
```
devsnek, line 927: more newline eslint rules to enable

rollup.config.js
Jack-Works: Oh, I thought engine262 needs to support some old browser (because I see it contains a transform for `?.` syntax).

Is it ok to use the latest syntax? (that means class initialization block comes in)

src/parser/FunctionParser.mts
Jack-Works: This is compiled using babel's TS support.

```ts
enum FunctionKind {
   Sync, Async,
}
```

goes to

```js
"use strict";

var FunctionKind;
(function (FunctionKind) {
  FunctionKind[FunctionKind["Sync"] = 0] = "Sync";
  FunctionKind[FunctionKind["Async"] = 1] = "Async";
})(FunctionKind || (FunctionKind = {}));
```

src/value.mts
Jack-Works: No idea. Does the `Type` in the spec accept any other type than an ECMAScript language value?

jmdyck: Not since https://github.com/tc39/ecma262/pull/2874 was merged, I think.

rollup.config.js
devsnek: i guess the main requirement is that it runs in node, which usually lags slightly behind browsers, so as long as it runs alright in node its usually fine.

src/value.mts
Jack-Works: Since https://github.com/tc39/ecma262/issues/2870 prefers Type(x) only accept ES language value, I will keep this signature, and if there is any new type error, it might indicate we find a something we need to change in the spec.

devsnek: regardless of the spec what I meant is that it doesn't match the code in the body of the function. I'm surprised ts didn't pick up on that somehow?

Jack-Works: image
yeah, typescript allows this 🤣

src/parser/FunctionParser.mts
Jack-Works: because arrowInfo also gets the type from `ParseNode`'s `[index: string]: any` signature. It might be improved in the future when ParseNode is detailed typed.

Jack-Works:
I added the eslint rules you requested, but looks like it doesn't lint the case you pointed out, strange.

I already change babel-env to use the last 2 node major versions.

Jack-Works:
I'll rebase this after #200 and other big changes land, @devsnek please cc me when this happens thanks!

Jack-Works:
I'll split them in many small PRs

PR #207, opened 2022-12-24
refactor: introduce typescript

Jack-Works:
meta: #202, reopen of #206

I did not include extra changes and I hope it can be merged once it's ok to reduce merge conflicts.

Note: Commit 8e1c6e2 should appear in the main history after merge. I use a [`.git-blame-ignore-revs`](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/) file to ignore the automatic change in this commit.

# Infra changes

Add TypeScript support in eslint and rollup config. Does not add type checker to the CI yet.

## Rename all files

1. Rename all files from .mjs to .mts
2. Add `// @ts-nocheck` at the top of the file

This allows me to write TypeScript syntax in any file but not be type checked.
When any file is ready for type checked, just remove the `// @ts-nocheck` comment.

## Babel target

last 2 node versions

## Class fields

TypeScript requires class fields syntax to declare a field. Using [[Define]].

This should be a breaking change for downstream users (if there is any).

## Decorator

I will use decorators to create callable classes.

```js
export @callable() class Descriptor { ... }
```

## Shipping declarations

Now after running `tsc -p .`, declarations/ folder contains the `.d.ts` file of engine262


Jack-Works:
@devsnek Hi can you review this? or maybe add me as a member of engine262 and I can migrate without bothering everyone with a lot of PR review requests 😇

Jack-Works:
thanks for inviting me to the org! I'll take that as an agreement of merging this!

Jack-Works:
Note: if anyone has problem with cannot run `npm run lint`, please `rm -rf node_modules/@engine262/eslint-plugin` and run `npm install` again

PR #211, opened 2023-02-27
Update modules graph evaluation algorithms

nicolo-ribaudo:
- Some fields have been renamed in ECMA-262
- The logic to track ancestors in AsyncModuleExecutionFulfilled has been updated to properly transition every module in cycles

This fixes a bug where the [[Status]] of some modules in cycles with TLA wasn't properly updated from `evaluating-async` to `evaluated`, even if it's not observable in user code.


nicolo-ribaudo:
@Jack-Works This PR aligns the module algorithms with the current ECMA-262 spec.

It also already includes https://github.com/tc39/ecma262/pull/3583, otherwise multiple test262 tests crash. I would leave it in even if it's not merged yet in ECMA-262, since it otherwise causes us to read some properties from `undefined`. I'll submit a PR updating the implementation if that spec PR will need to change.

Jack-Works:
hi does it also includes https://github.com/tc39/ecma262/pull/3535?

PR #223, opened 2023-07-10
Support Explicit Resource Management Proposal

rbuckton:
This adds support for the Explicit Resource Management Proposal, including both the sync and async forms.

This is a work in progress and is not yet ready for review.


Jack-Works:
hi! Are you still interested in this PR? This proposal is already in stage 4, and I think it should be the next priority of this project.

Jack-Works:
I'll take care of this! Close for now.

PR #224, opened 2023-07-20
Refactor completions into a discriminated union (alt)

rbuckton:
This is an alternative to #222 that is far less complicated, though it does mean more changes to existing code for the following cases:

- `x instanceof AbruptCompletion` to `isAbruptCompletion(x)`
- `x instanceof NormalCompletion` to `isNormalCompletion(x)`
- `x instanceof Completion` to `x instanceof CompletionRecord`
- `new Completion({ ... })` to `new CompletionRecord({ ... })`

### DIFF
--- a/bin/test262_realm.js
+++ b/bin/test262_realm.js
@@ -11,7 +11,7 @@ const {
   ToString,
   Type,
   Throw,
-  AbruptCompletion,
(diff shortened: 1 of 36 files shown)


Jack-Works:
close since #222 is merged

PR #227, opened 2023-08-17
Implement AsyncContext (Stage 2)

andreubotella:
This patch also updates some comments and code to better reflect the current spec.


src/abstract-ops/realms.mts
```diff
@@ -353,6 +364,7 @@ export function SetDefaultGlobalBindings(realmRec) {
     'WeakSet',
 
     // Other Properties of the Global Object
+    'AsyncContext',
```
Jack-Works, line 367: I think you should use a feature flag to conditionally expose this feature.

PR #229, opened 2023-11-15
Adds support for Stage 3 RegExp Modifiers

rbuckton:
This adds support for the Stage 3 [RegExp Modifiers](https://github.com/tc39/proposal-regexp-modifiers) proposal, which passes the tests in https://github.com/tc39/test262/pull/3960.

This also makes a few changes to several RegExp-related function names to match a more recent version of the ECMAScript specification.


Jack-Works:
Hi! Thanks for the PR, but #293 rewrites the whole RegExp and includes this feature.

PR #246, opened 2025-03-04
fix Promise.try

ajvincent:


ajvincent:
I would like to understand how the TypeScript compiler let me get away with this mistake the first time. :angry:

Jack-Works (approved):

Jack-Works:
> I would like to understand how the TypeScript compiler let me get away with this mistake the first time. 😠

we currently don't enable type checks because ts migration is not done.

PR #250, opened 2025-03-13
fixup! Implement "Symbols as WeakMap keys"

ExE-Boss:
No other part of the codebase does checking of what subtype a `Value` is by reading the `type` property.


ajvincent:
😅  sorry about this, still new to the code base.

PR #259, opened 2025-03-26
Implement proposal “Error Stack Accessor”

ExE-Boss:
Split off from <https://github.com/engine262/engine262/pull/157>.

## Spec:
- https://tc39.es/proposal-error-stack-accessor/

### DIFF
--- a/src/helpers.mts
+++ b/src/helpers.mts
@@ -13,7 +13,9 @@ import {
 } from './abstract-ops/all.mts';
 import { Q, X } from './completion.mts';
 import type { ParseNode } from './parser/ParseNode.mts';
-import type { Evaluator, EvaluatorNextType, YieldEvaluator } from './evaluator.mts';
+import type {
+  Evaluator, EvaluatorNextType, ValueEvaluator, YieldEvaluator,
+} from './evaluator.mts';
 import type { ErrorObject } from './intrinsics/Error.mts';
 
 export const kInternal = Symbol('kInternal');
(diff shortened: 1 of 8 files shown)


Jack-Works (changes requested):
Old stack setter is defined by `X(captureStack(O)); // NON-SPEC` but both old and new are present now

src/helpers.mts
```diff
@@ -530,41 +533,32 @@ export function getCurrentStack(excludeGlobalStack = true) {
...
-      }
-      return cache;
-    }, 0, name, [], undefined, undefined, Value('get')),
-    Set: CreateBuiltinFunction(([value = Value.undefined]) => {
-      cache = value;
-      return Value.undefined;
-    }, 1, name, [], undefined, undefined, Value('set')),
-    Enumerable: Value.false,
-    Configurable: Value.true,
-  })));
```
ExE-Boss, line 567: @Jack-Works
> Old stack setter is defined by X(captureStack(O)); // NON-SPEC but both old and new are present now

No, I deleted that here.

Jack-Works (approved):
LGTM, let me know if you think it is ready to merge. I wonder if there are corresponding test262 tests. I think there is no need to add a new FEATURE to keep the old behaviour.

ExE-Boss:
There are currently no **test262** tests as far as I’m aware.

Should be good to merge now.

Jack-Works (approved):

PR #265, opened 2025-04-02
Implement Iterator Helpers.

ajvincent:
This isn't ready yet.  Currently failing hundreds of tests, most likely due to some loose wires around registering the `Iterator` constructor.

Also, I'd really appreciate a close review, of the "this guy doesn't really know what he's doing yet" variety.  Because it's true:  I was following the specification as closely as I could, but that doesn't mean I have the deeper understanding of what this requires.

PR #266, opened 2025-04-02
Implement `Iterator` constructor

ExE-Boss:
Blocks <https://github.com/engine262/engine262/pull/265>.

Implementing this correctly required making changes to engine262’s prototype and constructor bootstrapping process to avoid overwriting the `Iterator.prototype.constructor` accessor, which is why I did this part myself.

--------------------------------------------------------------------------------

Unfortunately, test262 doesn’t have a feature flag for just testing the `Iterator` constructor on its own.

### DIFF
--- a/src/intrinsics/Iterator.mts
+++ b/src/intrinsics/Iterator.mts
@@ -1,0 +1,39 @@
+import {
+  OrdinaryCreateFromConstructor,
+  type BuiltinFunctionObject,
+  type Realm,
+} from '../abstract-ops/all.mts';
+import { Q, type ValueEvaluator } from '../completion.mts';
+import { surroundingAgent } from '../host-defined/engine.mts';
+import {
+  UndefinedValue,
+  type Arguments,
+  type FunctionCallContext,
+  type ObjectValue,
+} from '../value.mts';
+import { bootstrapConstructor } from './bootstrap.mts';
+
+/** https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-iterator-constructor */
+function* IteratorConstructor(
+  this: BuiltinFunctionObject,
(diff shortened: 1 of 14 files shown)


src/intrinsics/bootstrap.mts
```diff
@@ -132,12 +137,14 @@ export function bootstrapConstructor(realmRec: Realm, Constructor: NativeSteps,
     Configurable: Value.false,
   })));
 
-  X(Prototype.DefineOwnProperty(Value('constructor'), Descriptor({
-    Value: cons,
-    Writable: Value.true,
-    Enumerable: Value.false,
-    Configurable: Value.true,
-  })));
+  if (!Prototype.properties.has('constructor')) {
```
Jack-Works, line 140: Why did you add this guard?

src/intrinsics/Iterator.mts
```diff
@@ -0,0 +1,42 @@
...
+function* IteratorConstructor(
+  this: BuiltinFunctionObject,
+  _args: Arguments,
+  { NewTarget }: FunctionCallContext,
+): ValueEvaluator<ObjectValue> {
+  // 1. If NewTarget is either undefined or the active function object, throw a TypeError exception.
+  if (NewTarget === Value.undefined) {
+    return surroundingAgent.Throw('TypeError', 'ConstructorNonCallable', this);
+  }
+  __ts_cast__<FunctionObject>(NewTarget);
```
Jack-Works, line 28: You should use `NewTarget instanceof UndefinedValue` to avoid `__ts_cast__`. `__ts_cast__` is like `any` which is very type-unsafe.

Jack-Works (approved):

PR #269, opened 2025-04-04
Normative: Error.isError

Jack-Works:


src/intrinsics/Error.mts
```diff
@@ -70,8 +69,14 @@ function* ErrorConstructor([message = Value.undefined, options = Value.undefined
   return O;
 }
 
+function Error_isError([value]: Arguments) {
```
ExE-Boss, line 72: Missing specification link:
```suggestion
/** https://tc39.es/proposal-is-error/#sec-error.iserror */
function Error_isError([value]: Arguments) {
```

Issue #262, opened 2025-03-30
Implement iterator-helpers


ajvincent:
["Because it's _there_."](https://tc39.es/proposal-iterator-helpers)

I'm working on this.

PR #277, opened 2025-04-07
Implement iterator helpers

ajvincent:
So this is complete except for five tests failing:

- [staging/sm/Iterator/prototype/every/check-fn-after-getting-iterator.js](https://github.com/tc39/test262/blob/b3f67f4aafc9372ab7bfaea4290413b3e241f6a5/test/staging/sm/Iterator/prototype/find/check-fn-after-getting-iterator.js)
- staging/sm/Iterator/prototype/find/check-fn-after-getting-iterator.js
- staging/sm/Iterator/prototype/forEach/check-fn-after-getting-iterator.js
- staging/sm/Iterator/prototype/reduce/check-fn-after-getting-iterator.js
- staging/sm/Iterator/prototype/some/check-fn-after-getting-iterator.js

All five of these share the same pattern of failure: missing `get: return` from the proxy trap.  However, I looked into this and it looks like we're doing the right thing.  [A commit after our latest test262 update](https://github.com/tc39/test262/commit/d6897d560b73ec898aa12bd0739ead49cc9104aa) confirms this.


ajvincent:
Fixes #262 (and I didn't even notice I got the lucky number 🤣 )

src/intrinsics/Iterator.mts
```diff
@@ -31,8 +37,33 @@ function* IteratorConstructor(
...
 
+/** https://tc39.es/ecma262/#sec-iterator.from */
+function* Iterator_from([O]: Arguments): ValueEvaluator {
+  // 1. Let iteratorRecord be ? GetIteratorFlattenable(O, iterate-string-primitives).
+  const iteratorRecord = Q(yield* GetIteratorFlattenable(O, 'iterate-string-primitives'));
+
+  // 2. Let hasInstance be ? OrdinaryHasInstance(%Iterator%, iteratorRecord.[[Iterator]]).
+  const hasInstance = Q(yield* OrdinaryHasInstance(surroundingAgent.intrinsic('%Iterator%'), iteratorRecord.Iterator));
+  // 3. If hasInstance is true, then
+  if (hasInstance.booleanValue()) {
```
Jack-Works, line 48: in the codebase it usually should be `hasInstance === Value.true`, but this is not wrong either

ajvincent: fixed

PR #280, opened 2025-04-15
Add internal slot for inspecting closures.

ajvincent:
I discussed the premise of this with Jack-Works a couple days ago. Basically, there's a couple of reasons to add this slot:

1. I need it to reach the enclosed values of _built-in_ closures we establish in `Promise.finally()` or the iterators of `Array.prototype.values()`.
2. This may let the closures mark the enclosed values via `ObjectValue.prototype.mark()`, for the garbage collection algorithm.

I am deliberately excluding `CreateRegExpStringIterator` from this, as Jack-Works is actively working on engine262's regular expressions support.


src/abstract-ops/generator-operations.mts
```diff
@@ -33,6 +33,8 @@ export interface GeneratorObject extends OrdinaryObject {
   GeneratorContext: ExecutionContext | null;
   readonly GeneratorBrand: JSStringValue | undefined;
   UnderlyingIterator?: IteratorRecord;
+  // NON-SPEC
+  EnclosedValue?: Value;
```
Jack-Works, line 37: I'm thinking, should we put all captured values in all abstract closures in this field (by adding this field to including BuiltinFunctionObject)

So I'd suggest the name to be `HostCapturedValues?: readonly Value[];`. I'm not clear about the action we should take here (maybe we have more GC issues than this), so I will come back to this PR after I finish the RegExp update.

Jack-Works:
I revisited this PR and am wondering if we can have a better API, for example,

```
capture([a list of values that O has a strong reference to], O);
```

And what `capture` do is attach a `mark` method so it can work with GC

Jack-Works:
Oh, I read the conversation again. I hope you can rename it like `HostCapturedValues?: readonly Value[];` so it can (hopefully) scale to other host objects without introducing another API change.

ajvincent:
Although this is an internal slot I'm adding, I would like an automated test for this.  [`test-supplemental.mts`](https://github.com/engine262/engine262/blob/main/test/supplemental.mts) is getting pretty long though, and adding tests for this could make it quite a bit longer.

Jack-Works (approved):

PR #283, opened 2025-04-17
fix(scripts/transform): Emit supported c8 ignore comments

ExE-Boss:
**[c8](https://www.npmjs.com/package/c8)** doesn’t currently support `ignore if`, which is why the coverage was so abysmal, this migrates to using `node:coverage ignore next` and `node:coverage disable/enable`, which are supported by **c8** and **Node**’s experimental [coverage collector](https://nodejs.org/api/test.html#collecting-code-coverage).

This also makes it so that the <code>\`${name}.section = '${url}';\`</code> assignment happens before any trailing comments in case they contain coverage directives and adds `ignore next` comments to `throw new OutOfRange(…)` calls outside of the `default` `switch` clause, such as in: <https://github.com/engine262/engine262/blob/9bf790fe28af5147c9b600492d527b28ae16f384/src/runtime-semantics/Literal.mts#L18-L27>

### DIFF
--- a/scripts/transform.mts
+++ b/scripts/transform.mts
@@ -125,7 +125,11 @@ export default ({ types: t, template }: typeof import('@babel/core')): PluginObj
           if (/#sec/.test(line)) {
             const section = line.split(' ').find((l) => l.includes('#sec'))!;
             const url = section.includes('https') ? section : `https://tc39.es/ecma262/${section}`;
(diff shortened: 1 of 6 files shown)


Jack-Works:
the coverage dropped from 90+ to 46 after https://github.com/engine262/engine262/pull/261 merged. It looks like only code touched by `npm run test:inspector` is collected, and I have totally no idea why

PR #285, opened 2025-04-19
refactor(messages): Add types

ExE-Boss:
I also had to fix the definition of `Value[Symbol.hasInstance]`, which was incorrectly declared as an instance member instead of a static member.


src/messages.mts
```diff
@@ -1,48 +1,54 @@
...
-
-import { surroundingAgent, inspect, Value } from './index.mts';
+import type { ArrayBufferObject, FunctionObject } from './abstract-ops/all.mts';
+import type { AbstractModuleRecord } from './modules.mts';
+import type {
+  BooleanValue, JSStringValue, NumberValue, ObjectValue, PropertyKeyValue,
+} from './value.mts';
+import { inspect, PrivateName, Value } from './index.mts';
 
 function i(V: unknown) {
```
Jack-Works, line 8: I'm planning to get rid of `inspect` in error throwing, or improving it with a side-effect-free way of this. It used to cause test262 failure in https://github.com/engine262/engine262/pull/277/files#diff-e02b9e5bf3aa21bec763c8430087794b58a0f38acab9c15a3882825bf115ce2aL59

src/value.mts
```diff
@@ -848,6 +866,8 @@ export class ObjectValue extends Value implements ObjectInternalMethods<ObjectVa
   static {
     Object.defineProperty(this.prototype, 'type', { value: 'Object' });
   }
+
+  declare static [Symbol.hasInstance]: (value: unknown) => value is ObjectValue;
```
Jack-Works, line 870: This should be the default behaviour of `instanceof`, we don't need this signature to specify it, so I'm wondering if we can remove the whole `hasInstance` or not.

src/messages.mts
```diff
@@ -1,48 +1,54 @@
...
 export const CannotMixBigInts = () => 'Cannot mix BigInt and other types, use explicit conversions';
 export const CannotResolvePromiseWithItself = () => 'Cannot resolve a promise with itself';
-export const CannotSetProperty = (p, o) => `Cannot set property ${i(p)} on ${i(o)}`;
+export const CannotSetProperty = (p: PropertyKeyValue, o: unknown) => `Cannot set property ${i(p)} on ${i(o)}`;
 export const ClassMissingBindingIdentifier = () => 'Class declaration missing binding identifier';
 export const ConstDeclarationMissingInitializer = () => 'Missing initialization of const declaration';
-export const ConstructorNonCallable = (f) => `${i(f)} cannot be invoked without new`;
-export const CouldNotResolveModule = (s) => `Could not resolve module ${i(s)}`;
+export const ConstructorNonCallable = (f: Value) => `${i(f)} cannot be invoked without new`;
+export const CouldNotResolveModule = (s: JSStringValue | '') => `Could not resolve module ${i(s)}`;
```
Jack-Works, line 51: Many messages provide strings directly, so this type is not correct.

Jack-Works (changes requested):
I'm also thinking of rearranging error messages, like this, but I'm not sure.

``` js
E.ArrayBuffer.detached
```

Or even do we need a message file or not, since we're not translating it, and adding a new message is tedious. It's still worthy to type this file before we have any new idea of this.

src/value.mts
ExE-Boss: This is needed to override the `(value: unknown) => value is Value` declaration in `BaseValue`, otherwise `value instanceof ObjectValue` will break with the fixed `Symbol.hasInstance` method type in `BaseValue`.

src/messages.mts
ExE-Boss: Yeah, I plan to fix `inspect` to be side‑effect free (similar to <https://github.com/engine262/engine262/pull/117>, but better)

src/value.mts
Jack-Works: I see, it is because how we define Value

Jack-Works (approved):

PR #286, opened 2025-04-23
fix: Rewrite `inspect(…)` to be (probably) side‑effect free

ExE-Boss:
This rewrites `inspect(…)` in a way which avoids calling proxy traps and property accessors.
Other non‑`Proxy` implementations of the object internal methods `[[OwnPropertyKeys]]`, `[[GetOwnProperty]]`, and `[[GetPrototypeOf]]` seem to be side‑effect free, as far as I can tell.

In the process of doing this, I consolidated a lot of the code paths so that own properties are shown for all object types, including functions, primitive boxes, `Promise`es, `RegExp`s, and `Array`s.

Partially based on <https://github.com/engine262/engine262/pull/117>.

### DIFF
--- a/src/host-defined/inspect.mts
+++ b/src/host-defined/inspect.mts
@@ -9,20 +9,19 @@ import {
   Assert, EscapeRegExpPattern,
   IsAccessorDescriptor, isArrayExoticObject,
   isArrayIndex, IsCallable, IsDataDescriptor,
-  isECMAScriptFunctionObject,
+  isECMAScriptFunctionObject, IsError,
   LengthOfArrayLike, R, Realm,
   type FunctionObject,
 } from '../abstract-ops/all.mts';
 import { Completion, X, type ValueCompletion } from '../completion.mts';
 import { isRegExpObject } from '../intrinsics/RegExp.mts';
-import type { DateObject } from '../intrinsics/Date.mts';
-import type { BooleanObject } from '../intrinsics/Boolean.mts';
+import { isDateObject } from '../intrinsics/Date.mts';
+import { isBooleanObject } from '../intrinsics/Boolean.mts';
 import type { NumberObject } from '../intrinsics/Number.mts';
-import type { BigIntObject } from '../intrinsics/BigInt.mts';
+import { isBigIntObject } from '../intrinsics/BigInt.mts';
 import type { StringObject } from '../intrinsics/String.mts';
 import type { SymbolObject } from '../intrinsics/Symbol.mts';
-import type { ErrorObject } from '../intrinsics/Error.mts';
-import type { PromiseObject } from '../intrinsics/Promise.mts';
+import { isPromiseObject } from '../intrinsics/Promise.mts';
 import { isProxyExoticObject } from '../intrinsics/Proxy.mts';
 import { isTypedArrayObject } from '../intrinsics/TypedArray.mts';
 import { __ts_cast__, OutOfRange } from '../helpers.mts';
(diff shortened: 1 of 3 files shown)


src/host-defined/inspect.mts
```diff
@@ -1,229 +1,592 @@
...
+  if (IsCallable(v)) {
+    return getFunctionType(v);
+  }
+  if (isTypedArrayObject(v)) {
+    return v.TypedArrayName.stringValue();
+  }
+  if (isArrayExoticObject(v)) {
+    return 'Array';
+  }
+  if ('PromiseState' in v) {
```
Jack-Works, line 95: you can use helpers like `isPromiseObject`

src/host-defined/inspect.mts
```diff
@@ -1,229 +1,592 @@
...
+    case -1:
+      return `"${str}"`;
+
+    case 0x27:
+      return `'${str}'`;
+
+    default:
+      throw new OutOfRange(`strEscape: Invalid quote code point: ${quote}`, quote);
+  }
+};
```
Jack-Works, line 230: linter error

src/host-defined/internal-util.mts
```diff
@@ -0,0 +1,37 @@
...
+  type PropertyKeyValue,
+} from '../value.mts';
+
+/**
+ * Like `OrdinaryGet`, but doesn't call into proxies or invoke getters.
+ */
+export function* getNoSideEffects(
+  O: ObjectValue,
+  P: PropertyKeyValue,
+): PlainEvaluator<Value | (Descriptor & { Get: Value; Set: Value }) | undefined> {
```
Jack-Works, line 18: If the function is always called with `X`, the signature should be

```suggestion
export function getNoSideEffects(
  O: ObjectValue,
  P: PropertyKeyValue,
): Value | (Descriptor & { Get: Value; Set: Value }) | undefined {
```

src/host-defined/internal-util.mts
```diff
@@ -0,0 +1,37 @@
...
+  Assert(IsPropertyKey(P));
+
+  const desc = Q(yield* O.GetOwnProperty(P));
+  if (desc instanceof UndefinedValue) {
+    const parent = Q(yield* O.GetPrototypeOf());
+    if (parent instanceof NullValue || isProxyExoticObject(parent)) {
+      return undefined;
+    }
+    return Q(yield* getNoSideEffects(parent, P));
+  }
```
Jack-Works, line 29:
```suggestion
  const desc = X(O.GetOwnProperty(P));
  if (desc instanceof UndefinedValue) {
    const parent = X(O.GetPrototypeOf());
    if (parent instanceof NullValue || isProxyExoticObject(parent)) {
      return undefined;
    }
    return X(getNoSideEffects(parent, P));
  }
```

src/host-defined/inspect.mts
```diff
@@ -1,229 +1,592 @@
...
-      if (name && (name.Value! as JSStringValue).stringValue() !== '') {
-        return `[Function: ${(name.Value as JSStringValue).stringValue()}]`;
+      base = ctx.stylize(getFunctionBase(v as FunctionObject, constructor, tag), 'special');
+      if (keys.length === 0) {
+        return base;
       }
-      return '[Function]';
     }
 
     if ('ErrorData' in v) {
```
Jack-Works, line 409: use `isErrorObject` to avoid `(v as ErrorObject)`

Jack-Works (changes requested):

Jack-Works (changes requested):

src/host-defined/internal-util.mts
```diff
@@ -0,0 +1,36 @@
...
+  P: PropertyKeyValue,
+): Value | (Descriptor & { Get: Value; Set: Value }) | undefined {
+  Assert(O instanceof ObjectValue && !isProxyExoticObject(O));
+  Assert(IsPropertyKey(P));
+
+  const desc = X(O.GetOwnProperty(P));
+  if (desc instanceof UndefinedValue) {
+    const parent = X(O.GetPrototypeOf());
+    if (parent instanceof NullValue || isProxyExoticObject(parent)) {
+      return undefined;
```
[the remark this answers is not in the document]
ExE-Boss: This intentionally uses `undefined` (spec value `empty`) to disambiguate from a data property with a value of `Value.undefined`.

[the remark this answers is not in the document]
Jack-Works: As I understanding this case is accessing `x.a` where prototype of `x` is null, in this case, `x.a` should be `undefined` (language value), is that correct?

PR #291, opened 2025-05-05
Implement support for import attributes and JSON modules

nicolo-ribaudo:
This PR can be reviewed by commit:
- The first one implements import attributes
- The second one allows hosts to actually observe and use the import attributes (:warning: it includes a breaking change in the signature of `loadImportedModule`)
- The third one aligns the Synthetic Module Record implementation with the spec, and considers the `type: "json"` attribute in the built-in module loader (used for test262)


Jack-Works:
thanks!

PR #293, opened 2025-05-10
feat: rewrite RegExp

Jack-Works:
Implements:

- RegExp.escape
- RegExp modifiers
- RegExp v flag
- Duplicated named capture groups

passes all tests except the following one (and also, no Annex B)

- https://github.com/tc39/ecma262/issues/3592

This PR will be merged on May 24 if no reviews are pending at that time.


src/parser/Lexer.mts
```diff
@@ -25,12 +25,16 @@ export const isHexDigit = (c: string) => c && /[\da-f]/ui.test(c);
 const isOctalDigit = (c: string) => c && /[0-7]/u.test(c);
 const isBinaryDigit = (c: string) => (c === '0' || c === '1');
 export const isWhitespace = (c: string) => c && (/[\u0009\u000B\u000C\u0020\u00A0\uFEFF]/u.test(c) || isSpaceSeparatorRegex.test(c)); // eslint-disable-line no-control-regex
-export const isLineTerminator = (c: string) => !!c && /[\r\n\u2028\u2029]/u.test(c);
+export const isLineTerminator = (c: string | number) => {
+  if (typeof c === 'string') {
+    return !!c && /[\r\n\u2028\u2029]/u.test(c);
+  } else {
+    return c === 8232 || c === 8233 || c === 10 || c === 13;
```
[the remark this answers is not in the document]
ajvincent: Also, quite a long time ago I heard that else after return is a Bad Thing.  Not one that's ever afflicted JavaScript, but still.

src/parser/RegExpParser.mts
```diff
@@ -321,70 +401,152 @@ export class RegExpParser {
...
+  //
+  // RegularExpressionModifiers ::
+  //   [empty]
+  //   RegularExpressionModifiers RegularExpressionModifier
+  //
+  // RegularExpressionModifier :: one of `i` `m` `s`
+  private parseAtomModifiers(): {
+    PlusModifiers: ParseNode.RegExp.RegularExpressionModifier[] | undefined;
+    MinusModifiers: ParseNode.RegExp.RegularExpressionModifier[] | undefined;
+    } {
```
ajvincent, line 484: I'm not a big fan of inline object literal types as return types.

Suggestion:
```typescript
private parseAtomModifiers(): Record<"PlusModifiers" | "MinusModifiers", ParseNode.RegExp.RegularExpressionModifier[] | undefined> {
```

src/parser/RegExpParser.mts
```diff
@@ -855,11 +1089,17 @@ export class RegExpParser {
...
     for (let i = 0; i < length; i += 1) {
       const c = this.source[this.position];
       if (isHexDigit(c)) {
         this.position += 1;
+        oldN = n;
         n = (n << 4) | Number.parseInt(c, 16);
+        if (oldN > n) {
+          // overflow
+          this.raise('Invalid hex digit');
+        }
```
ajvincent, line 1102: On this I'm going to be somewhat ignorant, but I wonder if you've taken into account the sign bit.  `1 << 31` < 0.

src/parser/RegExpParser.mts
```diff
@@ -927,4 +1169,247 @@ export class RegExpParser {
...
+    }
+
+    // \ ClassSetReservedPunctuator
+    if ('& - ! # % , : ; < = > @ ` ~'.split(' ').includes(nextTwo[1])) {
+      this.position += 2;
+      return { type: 'ClassSetCharacter', production: 'UnicodeCharacter', UnicodeCharacter: nextTwo[1] as UnicodeCharacter };
+    }
+
+    // anything that can start a Character Escape
+    if ('f n r t v c 0 x u / ^ $ \\ . * + ? ( ) [ ] { } |'.split(' ').includes(nextTwo[1])) {
```
ajvincent, line 1343: I know engine262 is more concerned with correctness over efficiency.  But this function seems like it would benefit from having these long split strings be constants defined outside the method, say, as private static members.

src/runtime-semantics/RegExp.mts
```diff
@@ -1,1187 +1,1307 @@
...
-  captures;
+export type RegExpMatchingSource = (readonly string[]) & { readonly raw: string };
+/** https://tc39.es/ecma262/#pattern-matchstate */
+class MatchState {
+  readonly input: RegExpMatchingSource;
+
+  readonly endIndex: number;
 
-  constructor(endIndex: number, captures: (UndefinedValue | Range)[]) {
+  readonly captures;
```
ajvincent, line 36: missing type definition?

src/runtime-semantics/RegExp.mts
```diff
@@ -1,1187 +1,1307 @@
...
+    const setChars = sets.filter((x) => x.chars);
+    setChars.sort((a, b) => a.chars!.size - b.chars!.size)[0]?.chars?.forEach((c) => {
+      if (setChars.every((s) => s.chars!.has(c))) {
+        intersectionChars.add(c);
+      }
+    });
+
+    const intersectionStrings = new Set<ListOfCharacter>();
+    const setStrings = sets.filter((x) => x.strings);
+    setStrings.sort((a, b) => a.strings!.size - b.strings!.size)[0]?.strings?.forEach((c) => {
```
ajvincent, line 166: This is a little hard to read.  I think you're just looking for the smallest set size.  Using sort to do this is O(n log n), where a reduce might be O(n).

Of course, when n is small it doesn't really matter.  I'm more concerned about the readability of this code.

src/runtime-semantics/RegExp.mts
Jack-Works: TypeScript can infer the type of this field from the class constructor

src/parser/RegExpParser.mts
Jack-Works: I'm bad at bit calculation 😭, so I prefer the current code

ajvincent: Well, it won't be a problem unless length can be 8 or more.  There's only one place where that might be possible: `maybeParseRegExpUnicodeEscapeSequence` at line 1128.

There, you're going to throw for any sequence greater than `0x10FFFF`, which would be 6 characters.  So you can probably just throw the same exception:

```typescript
      if (end - this.position > 6) {
        this.raise('Invalid code point');
      }
```

ajvincent: Once you've done that, you can add an Assert call inside `scanHex` to make sure no one else breaks that rule in the future.

ajvincent: Heck, I could probably write an evil test262 case for this.

PR #294, opened 2025-05-21
Relax type definition of `Oridnary*` AOs to allow arbitrary objects

nicolo-ribaudo:
In the spec these methods take "an Object", and not "an ordinary object". They are indeed used on exotic objects, which might not have a [[Prototype]] or [[IsExtensible]] internal slot, so the current signature is wrong. I couldn't update the signature for the AOs related to PrototypeOf and IsExtensible, because they don't actually work on arbitrary objects even though the spec signature says so (see tc39/ecma262#3612).

Thanks to this change, we can remove some internal slot from the
`ModuleNamespaceObject` interface, aligning it with the spec.

### DIFF
--- a/src/abstract-ops/module-namespace-exotic-objects.mts
+++ b/src/abstract-ops/module-namespace-exotic-objects.mts
@@ -10,12 +10,10 @@ import {
   type ObjectInternalMethods,
   UndefinedValue,
   type PropertyKeyValue,
-  NullValue,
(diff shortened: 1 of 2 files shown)


Jack-Works (approved):

Jack-Works:
oh but if this change is not accepted in the spec, we may revert it in the future

PR #295, opened 2025-05-21
Implement `import defer`

nicolo-ribaudo:
Marking as draft because it depends on https://github.com/engine262/engine262/pull/294, and because there are 5 test262 failures I need to debug.


nicolo-ribaudo:
Thanks for merging the other PRs, this is ready now!

src/abstract-ops/module-namespace-exotic-objects.mts
```diff
@@ -31,11 +32,15 @@ import {
...
   type ExoticObject,
+  EvaluateModuleSync,
+  GetImportedModule,
 } from './all.mts';
+import type { ModuleRecord, PlainEvaluator } from '#self';
 
 export interface ModuleNamespaceObject extends ExoticObject {
   readonly Module: AbstractModuleRecord;
   readonly Exports: JSStringSet;
+  /* [import-defer] */ readonly Deferred: BooleanValue;
```
Jack-Works, line 43: the old convension is to use `BooleanValue`, but if the value is never mixed with a spec value, I would prefer to use `boolean` to improve interact with ts and also easier to inspect when debugging.

src/parser/ModuleParser.mts
```diff
@@ -20,7 +20,22 @@ export abstract class ModuleParser extends StatementParser {
     if (this.test(Token.STRING)) {
       node.ModuleSpecifier = this.parsePrimaryExpression();
     } else {
-      node.ImportClause = this.parseImportClause();
+      const importClause = this.parseImportClause();
+      if (
+        this.feature('import-defer')
+        && importClause.ImportedDefaultBinding?.sourceText() === 'defer'
```
Jack-Works, line 26: I'm not sure I understanded this part, does it means it parses like `import defer ` as part of `import defer from ...` (import default) and repurpose it as import phase? If so it looks hacky, it will be better if you can `this.peek()` to branch the parser.

src/abstract-ops/module-namespace-exotic-objects.mts
nicolo-ribaudo: I 100% agree with this and I was a bit annoyed by having to use BooleanValue, I'll change it :)

src/parser/ModuleParser.mts
nicolo-ribaudo: Yeah I can do `.peek()`, I was used to what we do in Babel's parser where we try to avoid peeking for performance reasons.

nicolo-ribaudo:
Fyi, this PR should now be ready :)

PR #296, opened 2025-05-22
Fix usage of wrong variable

nicolo-ribaudo:
I'm going to add a test262 test that would have caught this.

Spec: step 11.c.iv.3 of https://tc39.es/ecma262/#sec-innermoduleevaluation


nicolo-ribaudo:
Test: https://github.com/engine262/engine262/pull/295

PR #303, opened 2025-06-19
Replace [[DFSIndex]] with a local variable

nicolo-ribaudo:
Changes for https://github.com/tc39/ecma262/pull/3625, to make sure it doesn't break anything.

Marking as draft until when that PR is accepted :)


nicolo-ribaudo:
Spec PR merged!

PR #308, opened 2025-07-16
Implement set-methods (remaining parts)

ajvincent:


src/intrinsics/SetPrototype.mts
```diff
@@ -355,3 +709,38 @@ function SetDataHas(resultSetData: (Value | undefined)[], value: Value) {
   // 2. Return false.
   return Value.false;
 }
```
Jack-Works, line 711: I suggest change `SetDataHas` to return a `boolean` instead of a `BooleanValue` since it's return value is never used as a `Value`.

src/intrinsics/SetPrototype.mts
```diff
@@ -317,7 +671,7 @@ function* GetSetRecord(obj: Value): PlainEvaluator<SetRecord> {
 /** https://tc39.es/proposal-set-methods/#sec-getkeysiterator */
 function* GetKeysIterator(setRec: SetRecord): PlainEvaluator<IteratorRecord> {
```
Jack-Works, line 672: Hi, I cannot find this AO in the formal spec, would you like to investigate what happened? Thanks!

Jack-Works (changes requested):
Hi! Thanks for your PR. There are some code styles I hope you can change. Let me know if you want me to do it!

1. You should not write `x.value` or `x.value === false`, but `x === Value.true` or `x === Value.false`
2. I'd prefer `while` loops over `for` loops since it matches the spec.

ajvincent:
> Let me know if you want me to do it!

No, I can take care of these changes.  Though it would be nice to have linter rules catch this for us.  😄

src/intrinsics/SetPrototype.mts
```diff
@@ -163,17 +247,280 @@ function SetProto_sizeGetter(_args: Arguments, { thisValue }: FunctionCallContex
...
+        4. Set thisSize to the number of elements in O.[[SetData]].
+    */
+    let thisSize = O.SetData.length;
+    let index = 0;
+    while (index < thisSize) {
+      const e: Value | undefined = O.SetData[index];
+      index += 1;
+      if (e !== undefined) {
+        const inOther = ToBoolean(Q(yield* Call(otherRec.Has, otherRec.SetObject, [e])));
+        if (inOther === Value.true && SetDataHas(resultSetData, e) === false) {
```
Jack-Works, line 292:
```suggestion
        if (inOther === Value.true && !SetDataHas(resultSetData, e)) {
```

In this case, you can write normal JavaScript.

src/intrinsics/SetPrototype.mts
```diff
@@ -163,17 +247,280 @@ function SetProto_sizeGetter(_args: Arguments, { thisValue }: FunctionCallContex
...
+        1. If SetDataHas(O.[[SetData]], next) is true, then
+          a. Perform ? IteratorClose(keysIter, NormalCompletion(unused)).
+          b. Return false.
+    */
+    const keysIter = Q(yield* GetIteratorFromMethod(otherRec.SetObject, otherRec.Keys));
+    let next: Value | 'done' | 'not-started' = 'not-started';
+    while (next !== 'done') {
+      next = Q(yield* IteratorStepValue(keysIter));
+      if (next !== 'done' && SetDataHas(O.SetData, next)) {
+        Q(yield* IteratorClose(keysIter, NormalCompletion('unused')));
```
Jack-Works, line 390: `~unused~` should be `undefined`, not a string.

PR #309, opened 2025-07-17
Implement Array.fromAsync

ajvincent:
This does not quite work:

```
> Array.fromAsync([0, 1, 2])
Uncaught AssertError: completion.type === 'await-resume'
    at Assert (/Users/ajvincent/engine262/engine262/src/abstract-ops/notational-conventions.mts:12:11)
    at Await (/Users/ajvincent/engine262/engine262/src/completion.mts:480:3)
    at Await.next (<anonymous>)
    at ObjectValue.Array_fromAsync (/Users/ajvincent/engine262/engine262/src/intrinsics/Array.mts:280:29)
    at Array_fromAsync.next (<anonymous>)
    at BuiltinCallOrConstruct (/Users/ajvincent/engine262/engine262/src/abstract-ops/function-operations.mts:545:5)
    at BuiltinCallOrConstruct.next (<anonymous>)
    at Call (/Users/ajvincent/engine262/engine262/src/abstract-ops/object-operations.mts:203:38)
    at Call.next (<anonymous>) {
  [cause]: undefined
```

At the point of assertion failure, `completion-type` is "debugger-resume".  The step I fail at is 9.d.iv of [the `Array.fromAsync` proposal spec](https://tc39.es/proposal-array-from-async/)... line 282 of Array.mts in this patch.

Once again, I don't know if the bug is something I've done wrong or in the existing code.

Also, since I'm introducing the `IfAbruptCloseAsyncIterator` abstract operation, I would recommend taking a very close look at that.  I made an educated guess based on nearby code on its macro implementation.

### DIFF
--- a/scripts/transform.mts
+++ b/scripts/transform.mts
@@ -23,7 +23,7 @@ function getEnclosingConditionalExpression(path: NodePath) {
   return null;
 }
 
-type NeededNames = 'Completion' | 'AbruptCompletion' | 'Assert' | 'Call' | 'IteratorClose' | 'Value' | 'skipDebugger';
+type NeededNames = 'Completion' | 'AbruptCompletion' | 'Assert' | 'Call' | 'IteratorClose' | 'AsyncIteratorClose' | 'Value' | 'skipDebugger';
 
 interface State extends PluginPass {
   needed: Partial<Record<NeededNames, boolean>>;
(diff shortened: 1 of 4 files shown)


ajvincent:
I don't doubt there are other failures that I haven't diagnosed.  I was just looking for help with this assert failure, so I can move on to fixing what I can.

Jack-Works:
now it passes ✅ 158 and ❌ 13

ajvincent:
Down to one!

```
FAILED built-ins/Array/fromAsync/this-constructor.js
  Constructs the this-value once if asyncItems is iterable, twice if not, and length and element properties are set correctly on the result
```

I'll see if I can figure this out over the next day or so.  Smells like `thisArg` isn't what it should be.

Jack-Works (changes requested):

PR #330, opened 2025-12-30
Implement Promise.allKeyed, Promise.allSettledKeyed.

ajvincent:
Smoketest on `npm run inspector`:

```javascript
const blocking = Promise.withResolvers();
async function getShape() {
  await blocking.promise;
  return "cube";
}

async function getColor() {
  await blocking.promise;
  throw new Error("no color available");
}

async function getMass() {
  await blocking.promise;
  return 4000; // kg
}

const promiseDictionary = {
  shape: getShape(),
  color: getColor(),
  mass: getMass(),
};

const dictionary = Promise.allSettledKeyed(promiseDictionary);
blocking.resolve();

try {
  console.log(await dictionary);
} catch (ex) {
  console.log("caught exception");
  console.error(ex);
}
```

I was getting reasonable results when I tweaked this for all promises resolving, and for allSettled instead of allSettledKeyed.

As directed, this is behind a feature flag.  Nice that `npm run inspector` picked up on it automatically.


src/intrinsics/Promise.mts
```diff
@@ -212,6 +213,189 @@ function* Promise_all([iterable = Value.undefined]: Arguments, { thisValue }: Fu
   return result;
 }
 
+/** https://tc39.es/proposal-await-dictionary/#sec-promise.allkeyed */
+function* Promise_allKeyed([promises = Value.undefined]: Arguments, { thisValue }: FunctionCallContext): ValueEvaluator {
+  // 1. Let C be the this value.
```
Jack-Works: Hi! I would prefer to remove spec steps if they're very straitforward.

For example I will keep comments like this:

```js
// 1. If an implementation-defined debugging facility is available and enabled, then
```

```js
  // 1. If StringToCodePoints(text) is not a valid JSON text as specified in ECMA-404, throw a SyntaxError exception.
  Q(JSONValidator.validate(text));
```

Jack-Works:
Hi, I just pushed a big commit, if you have difficulties with merging plz let me know

Jack-Works:
Hi! Please revert your changes on `website` and `test262` submodule (unless it is necessary).

PR #346, opened 2026-05-05
Implement `export defer`

caiolima:
This implements `export defer` feature from https://github.com/tc39/proposal-deferred-reexports. Tests from Test262 are still in progress, and we should just land this PR once those tests lands on repository.


Jack-Works:
Hi! Do you want to merge it soon so other delegates can test it in playground/repl?

caiolima:
> Hi! Do you want to merge it soon so other delegates can test it in playground/repl?

Hi @Jack-Works, my worry here is just the coverage of tests we have so far. I estimate that I'll finish tests for this [test plan](https://github.com/tc39/test262/issues/5010) by middle of June, but I don't have idea on when they are going to be reviewed. If we are fine to have this as quite experimental, I'm down to merge it soon.

src/static-semantics/ModuleRequests.mts
```diff
@@ -48,6 +63,59 @@ function WithClauseToAttributes(node: ParseNode.WithClause): ImportAttributeReco
...
 
+/** Compute [[ImportedNames]] for a ModuleRequest derived from an ImportDeclaration's ImportClause.
+ * Per proposal-deferred-reexports:
+ * - `import "m"` (no clause) → « »
+ * - `import * as ns from "m"` (NameSpaceImport) → ~all~
+ * - `import a from "m"` (DefaultBinding) → « "default" »
+ * - `import { x, y as z } from "m"` (NamedImports) → « "x", "y" »
+ * - Combinations are unioned.
+ */
+function importedNamesFromImportClause(importClause: ParseNode.ImportClause | undefined): ImportedNamesValue {
```
Jack-Works, line 74: Does this map to an AO in the spec?

Jack-Works (approved):

src/static-semantics/ModuleRequests.mts
caiolima: It maps to [ImportNames](https://tc39.es/proposal-deferred-reexports/#sec-ImportedNames) SDO. Probably I should point there instead, right?

Jack-Works: yes! I'd suggest adding a link, and renaming it to `ImportedNames_FromImportClause`, we use upper cases for functions from spec (because in spec they're upper cases)

caiolima:
Thanks a lot for the review.

PR #348, opened 2026-05-21
[export-defer] Fixing eager evaluation of deferred re-exports on dynamic imports

caiolima:
Current specification is skipping async dependencies that are reachable from a sub-graph under a deferred re-exports, and this commit changes ContinueDynamicImport to also gather async dependencies for that part of the dependency graph.

It's a mirror of spec fix https://github.com/tc39/proposal-deferred-reexports/pull/38, and also fixes the eager evaluation happening on current implementation.

### DIFF
--- a/src/abstract-ops/module-records.mts
+++ b/src/abstract-ops/module-records.mts
@@ -718,7 +718,7 @@ export function GatherAsynchronousTransitiveDependenciesForRequests(
   const result: ModuleRecord[] = [];
   // 2. For each ModuleRequest Record request of moduleRequests, do
   for (const request of requests) {
-    if (request.Phase == 'source') {
+    if (request.Phase === 'source') {
       continue;
     }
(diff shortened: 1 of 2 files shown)


src/abstract-ops/module-records.mts
```diff
@@ -660,25 +660,8 @@ export function GatherAsynchronousTransitiveDependencies(module: ModuleRecord, s
...
-      // i. If result does not contain m, append m to result.
-      if (!result.includes(m)) {
-        result.push(m);
-      }
-    }
-  }
-  // 9. Return result.
-  return result;
+  // 8. Return GatherAsynchronousTransitiveDependenciesForRequests(module, module.[[RequestedModules]], seen)
+  return GatherAsynchronousTransitiveDependenciesForRequests(module, module.RequestedModules, seen);
```
caiolima, line 664: This was an overlook on initial commit that didn't change this AO properly to match current `export-defer` spec text.

PR #352, opened 2026-06-15
Rewrite evaluateScript, evaluate, jobQueue and add event loop

Jack-Works:
# Breaking changes

## Behavior of `realm.evaluateScript` or `realm.evaluate`

If you are using `realm.evaluateScript` or `realm.evaluate`, these APIs no longer require calling `resumeEvaluate()`.

`realm.evaluateScript` and `realm.evaluate` no longer return a completion.

```javascript
// old
let completion;
realm.evaluateScript('code', options, (c) => {
  completion = c;
  resolve(c);
});
if (!completion) surroundingAgent.resumeEvaluate({});

// new
realm.evaluateScript('code', options, (completion) => {
  resolve(completion);
});
```

If you want to preserve the old behavior, where you add an evaluator without running it, do this:

```javascript
realm.evaluateScript('code', options, (completion) => {
  resolve(completion);
}, false);
```

If you want to skip the debugger, do this:

```javascript
realm.evaluateScriptSkipDebugger('code', options);
// or
realm.evaluateScript('code', options, (completion) => {
  resolve(completion);
}, false, { noBreakpoint: true });
```

## Event loop & Job queue

`runJobQueue` has been removed. Use `agent.eventLoop` and `agent.jobQueue` instead. It by default is `WebLikeEventLoop` and will automatically flush.

### JobQueue API

```ts
export declare function runSingleJobInQueue(job: Job, onError: (error: Value) => void, finished: () => void): void;
export interface JobQueue extends Markable {
    enqueueFinalizationRegistryCleanupJob(job: Job): void;
    enqueuePromiseJob(job: Job): void;
    enqueueTimeoutJob(job: Job): void;
    enqueueGenericJob(job: Job): void;
    onNewJob: Set<(job: Job) => void>;
    shift(): Job | undefined;
    shiftFinalizationRegistryCleanupJob?(): Job | undefined;
    shiftPromiseJob?(): Job | undefined;
    shiftTimeoutJob?(): Job | undefined;
    shiftGenericJob?(): Job | undefined;
    get length(): number;
}
export declare class BasicJobQueue extends Set<Job> implements JobQueue, Markable {}
export declare class ByTypeJobQueue implements JobQueue, Markable {}
```

### EventLoop API

```ts
/** https://nodejs.org/learn/asynchronous-work/event-loop-timers-and-nexttick */
export type NodeJSJobType = 'timers' | 'pending callbacks' | 'idle' | 'prepare' | 'poll' | 'check' | 'close callbacks';
export type EventLoopRunType = 'manual' | 'automatic';
export interface EventLoop extends Markable {
    /**
     * Enqueue a host job (macrotask) now.
     */
    enqueue(type: NodeJSJobType | string, job: Job): void;
    /**
     * Enqueue a host job (macrotask) in the future.
     * It will keep `hasPendingJobs` to be `true` until `enqueue` is called.
     */
    enqueueAsync(type: NodeJSJobType | string, job: Job, executor: (enqueue: () => void, cancel: () => void) => void): void;
    /** Change the automatic run type of the event loop and run it */
    run(type: EventLoopRunType): void;
    runOnce(): void;
    readonly hasPendingJobs: boolean;
    /**
     * Register a callback to be called when there are no async queued jobs.
     *
     * This can be used to implement a mechanism to exit the program when all code has finished executing.
     */
    onNoPendingJob: Set<() => void>;
}

export declare abstract class AbstractEventLoop implements EventLoop {}

export declare class MicroTaskEventLoop extends AbstractEventLoop {}

/** https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide/In_depth */
export declare class WebLikeEventLoop extends AbstractEventLoop {}

/** https://nodejs.org/learn/asynchronous-work/event-loop-timers-and-nexttick */
export declare class NodeJSLikeEventLoop extends AbstractEventLoop {
    fallbackType: NodeJSJobType;
}
```

### Examples: timer

```ts
const job: Job = {
  queueName: 'timer',
  callerRealm: surroundingAgent.currentRealmRecord,
  callerScriptOrModule: GetActiveScriptOrModule(),
  job: function* setTimeoutResolve() {
    return yield* Call(callback, Value.undefined, []);
  },
};
// enqueueAsync prevents the event loop from exiting.
surroundingAgent.eventLoop.enqueueAsync('timers', job, (enqueue) => {
  // enqueue() actually enqueues the job.
  setTimeout(enqueue, ms.value);
});
```

## Removal of `realm.active` and `realm.scope()`

`realm.active` has been removed. It did not represent anything useful.

`realm.scope()` has been removed. Use `realm.pushTopContext()` instead.

```js
// old
realm.scope(() => {
  doSomething();
});

// new
const pop = realm.pushTopContext();
doSomething();
pop?.();
// in case of manually push and pop the context,
// you may need to run event loop manually.
agent.eventLoop.runOnce();
```

## `evalQ`

In `evalQ((Q, X) => {})`, the `X` parameter is no longer provided. Use `unwrapCompletion()` or `X()` instead.

## Renames

1. `ExecutionContext.codeEvaluationState` has been renamed to `ExecutionContext.CodeEvaluationState` to match ECMA-262 naming.

1. `Agent.scheduledForCleanup` has been renamed to `Agent.finalizationRegistryScheduledForCleanup`.

1. The `Agent.hostDefined.cleanupFinalizationRegistry` hook has moved to `Agent.hostDefinedOptions.hostHooks.HostEnqueueFinalizationRegistryCleanupJob`.

1. The type of `PrivateEnvironment` has changed from `PrivateEnvironmentRecord | NullValue` to `PrivateEnvironmentRecord`.

# Additions

1. Add `runningExecutionContext()` for the `running execution context` term. `surroundingAgent.runningExecutionContext` is still supported.
1. Add `currentRealmRecord()` for the `current realm record` term. `surroundingAgent.currentRealmRecord` is still supported.
1. Add `activeFunctionObject()` for the `active function object` term. `surroundingAgent.activeFunctionObject` is still supported.
1. Add `intrinsics()['%...%']` for intrinsics. `surroundingAgent.intrinsic('%...%')` is still supported, but discouraged.
1. `ModuleCache.hasUnfinishedRequests()` and `ModuleCache.untilAllRequestFinished()` have been removed.


Jack-Works:
@copilot resolve the merge conflicts in this pull request

lib-src/inspector/objects/modules.mts
```diff
@@ -11,51 +11,49 @@ export const Module = new ObjectInspector<ModuleNamespaceObject>('Module', undef
...
+        } else {
+          const realm = module.Module.Realm as ManagedRealm;
+          const evaluate = CreateBuiltinFunction(function* evaluate() {
+            return yield* (Get(module, key));
+          }, 0, 'Module.evaluate', [], realm);
+          result.push({
+            name: key.stringValue(),
+            get: getInspector(evaluate).toRemoteObject(evaluate, getObjectId, context, generatePreview),
+            set: { type: 'undefined' },
+            writable: false,
```
ajvincent, line 49: If these are meant to be property descriptors, or at least similar to them, we should try to treat `writable` as mutually exclusive with `get()` and `set()`.

lib-src/inspector/evaluator.mts
```diff
@@ -3,14 +3,17 @@ import {
   isEvaluator, ManagedRealm, skipDebugger, surroundingAgent, type PlainCompletion, type PlainEvaluator,
 } from '#self';
 
-export function nativeEvalInAnyRealm<T>(closure: (() => PlainCompletion<T>) | (() => PlainEvaluator<T>), context: InspectorContext): PlainCompletion<T> | undefined {
-  const realm = (surroundingAgent.runningExecutionContext?.Realm || context.getAnyRealm()?.realm) as ManagedRealm | undefined;
+export function nativeEvalInAnyRealm<T>(enterPreview: boolean, context: InspectorContext, closure: (() => PlainCompletion<T>) | (() => PlainEvaluator<T>)): PlainCompletion<T> | undefined {
```
ajvincent, line 6: This function feels important enough, based on what I've read above, to have some JSDoc or TSDoc comments.

`closure: (() => (PlainCompletion<T> | PlainEvaluator<T>))` ?

lib-src/inspector/evaluator.mts
```diff
@@ -3,14 +3,17 @@ import {
...
-  });
+
+  const exitPreview = enterPreview ? surroundingAgent.debugger_scopePreview() : undefined;
+  const pop = realm.pushTopContext();
+  let result = closure();
+  if (isEvaluator(result)) {
+    result = skipDebugger(result);
+  }
+  pop?.();
+  exitPreview?.[Symbol.dispose]();
```
ajvincent, line 17: I will note `Symbol.dispose` is not supported on Safari, according to [developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/dispose#browser_compatibility), and that's [the second most popular browser](https://gs.statcounter.com/) out there.

So this could break the playground for Safari users.

src/execution-context/Agent.mts
```diff
@@ -103,30 +123,36 @@ export class Agent {
...
+   * An event that is triggered when evaluators are finished and the agent is idle.
+   * @internal
+   */
+  readonly onNoEvaluator = new Set<() => void>();
 
   // NON-SPEC
-  /** This function will synchronously return a completion if this is a nested evaluation and debugger cannot be triggered. */
-  evaluate<T extends Value>(evaluator: ValueEvaluator<T>, onFinished: (completion: NormalCompletion<T> | ThrowCompletion) => void) {
+  /** Evaluate an evaluator. It will skip the debugger if the agent is already debugger-paused. */
+  evaluate<T extends Value>(evaluator: ValueEvaluator<T>, onFinished: (completion: NormalCompletion<T> | ThrowCompletion) => void, evaluationOptions?: ResumeEvaluateOptions | false): void {
```
ajvincent, line 145: Would it be a good idea to `Assert(!this.#evaluateFinishedCallback)`?

src/execution-context/ExecutionContext.mts
```diff
@@ -72,6 +93,23 @@ export class ExecutionContext {
...
+export class ExecutionContextStack extends Array<ExecutionContext> {
+  // This ensures that only the length taking overload is supported.
+  // This is necessary to support `ArraySpeciesCreate`, which invokes
+  // the constructor with argument `length`:
+  constructor(length = 0) {
+    super(+length);
+  }
+
+  // @ts-expect-error
+  override pop(ctx: ExecutionContext) {
```
ajvincent, line 105: What error?  Usually when I use `@ts-expect-error`, my linter complains that I need to explain something.

src/host-defined/event-loop.mts
```diff
@@ -0,0 +1,313 @@
...
+      marker(job.callerScriptOrModule);
+    }
+  }
+}
+
+const NODEJS_PHASES = [
+  'timers',
+  'pending callbacks',
+  'idle',
+  'prepare',
```
ajvincent: In the spec you reference, "idle" and "prepare" are at the same level... could there be a situation where an "idle" job is added while in the "prepare" phase, and we don't run it until we cycle all the way around again?

lib-src/inspector/objects/modules.mts
Jack-Works: The type definition is from Chrome Devtools protocol, I cannot change the type definition from npm.

lib-src/inspector/evaluator.mts
Jack-Works: This is not public API, only used in the inspector, you can PR to change it when needed, but this is usually run for some helper code that should never trigger the debugger.

lib-src/inspector/evaluator.mts
Jack-Works: For browsers that does not `Symbol.dispose`, it will become a string property `undefined` on both definition side and call site side `const disposbale = { undefined: () => {} }`. We only need to take care of this after we use the second Symbol-based proposal.

src/execution-context/ExecutionContext.mts
Jack-Works: This signature is incompatible with the base class Array.pop which does not require a parameter. This code is very old, I just move it to another file.

